PR #2, opened 2016-09-13
ECIP-1010

splix:
Delay Difficulty Bomb Draft


elaineo:
The file and explanations all make sense, although I prefer the "delay the bomb" strategy. When you "freeze", there is a huge jump in difficulty:
- Block 5,000,000 == 2**48 == 281 TH
- Block 5,200,000 == 2**50 == 1126 TH

Whereas with the "delay the bomb", the difficulty increases more gradually.

PR #9, opened 2016-10-01
Add a ECIP for adding blockchain origin to transactions.

marcusrbrown:
_DO NOT MERGE THIS PULL REQUEST UNTIL IT IS ACCEPTED AS A DRAFT_

This is based off of the discussion towards the end of ethereum/EIPs#134. By adding a Origin Block Hash field to the transaction header, clients can specify the origin of a transaction which assures it can only be valid on one or more blocks.

As noted in the Ethereum discussion, clients will need to maintain a list of block numbers of current and planned fork blocks, to be able to compare the `To` hash. Specifying a null hash (useful for hardware or offline wallets) will assure that the transaction is valid on any compatible chain but will not offer replay protection.

### DIFF
--- a/ECIPs/ECIP-1011.md
+++ b/ECIPs/ECIP-1011.md
@@ -1,0 +1,39 @@
+### Title
+
+    ECIP: 1011
+    Title: Transaction Replay Prevention
+    Status: Draft
+    Type: Standard
+    Layer: Consensus (hard-fork)
+    Author: Marcus R. Brown <marcus@ethereumclassic.org>
+    Created: 2016-09-25
+
+### Abstract
+This ECIP describes a strategy for preventing transaction replays (or "replay attacks") both on the blockchain and across forks. It adds a new field to the transaction header that specifies the origin of the blockchain where the transaction should be considered valid.
+
+### Motivation
+When a blockchain is created or a hard fork of an existing blockchain occurs, a distinct block is confirmed on the blockchain that represents either the "Genesis" (creator block) or fork point. Transactions created on blockchains (or a fork of an existing blockchain) that share an identical consensus algorithm can be replayed across both blockchains or forks under the following conditions:
+
+- The address of the account that created the transaction is the same on both blockchains.
+- The balance of the account that created the transaction contains enough Ether for the transaction to be valid on both blockchains.
+- The nonce of the account that created the transaction is lower or equal to the nonce of the account on another blockchain.
+
+In the case of a fork of an existing blockchain, it is usually undesirable to replay transactions occurring on the pristine blockchain onto the forked blockchain. In the case of similar but separate blockchains (e.g. a "testnet" versus a "mainnet"), it is usually undesirable to replay a transaction originating on one blockchain onto another.
+
+### Specification
+A new field is added to the transaction header (called Origin Block Hash, formally `To`) that contains the hash of a block on a blockchain where the transaction originates. The value of the `To` can be any one of the following:
+
+- A null hash, consisting of all zeros
+- The hash of the Genesis block
+- The hash of the most recent fork block
+
+A null hash is permitted as the value for `To` for transactions that are generated and signed offline. It can be assumed that any transaction where `To` contains the null hash is not safe from transaction replays across another blockchain for fork.
+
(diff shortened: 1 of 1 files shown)


elaineo:
Thanks Marcus! Just wondering: Is this any different from just using, say, a version number in the header?
"A new field is added to the transaction header (called Origin Block Hash, formally `To`"

marcusrbrown:
Yes, because as described in the ECIP, there may be situations where you want an offline transaction to always be valid on a given chain (whether it forks or not), or to lock it to a specific chain, or to a specific fork. I think you would need multiple interpretations of a version field in those case, also this unequivocally gives you the origin chain of a transaction.

I think the most contentious case would be the null hash since it has no replay protection, but that would need to be handled between the client and the user.

elaineo:
Okay, I understand. Bitcoin uses [version bits](https://gist.github.com/sipa/bf69659f43e763540550) to check for compatibility between blocks. Could the header hash be checked using a masking function such as `(tx.nVersion & ~expectedVersion) != 0` ?

I think that could allow for compatibility for future-dated tx's while still preserving replay protection.

splix:
I suggest to replace _"hash of the most recent fork block"_ with just _"hash of the most recent stable block"_. Because clients most likely don't know about _forks_, it's just a block in current chain

elaineo:
@igetgames @avtarsehra I've been outlining the implementation for this one, and it looks like for offline transactions, the user will need to specify additional options via console or command line. So it could be pretty involved to support that option. 

The replay attack fix is actually a soft fork. Should we include it in a future release instead of this one, in the interest of time?

marcusrbrown:
I was anticipating clients specifically built for offline transactions, such as CEW or hardware wallets, to use the null hash (offline) option. I don't think you would need that functionality in Geth, if that was your concern?

elaineo:
I'm thinking of sequencing transactions with a higher nonce. Like, I could
give you a "future-dated" signed tx with a higher nonce, and it won't be
compatible in the future if the block hash changes. I believe this is
required for state channels, but am not 100% sure.

marcusrbrown:
Closing in favor of #10.

elaineo:
I think it's still worthwhile to mention this, and the transaction-nonce-increase, to ethcore. They have probably spent more time looking into replay-attack-protection and may have pros/cons for the different strategies that we have not considered.

marcusrbrown:
Reopening on @elaineo's recommendation and to polish a bit.

avtarsehra:
@igetgames So has a decision been made on which option is the most favourable? I saw in the other PR that you wanted me to review something and agree to it. I wasn't sure which option for relay attack you were referring to. I am ok with all three options i.e. 1. Nonce offset, 2. Extra Field or 3. Mod to Signing Process. I think all have their pros and cons, obviously I have more interest in one that has minimal impact to transaction structure, but they would all equally work and solve the problem. So I would be happy to go with one that meets a consensus from everyone. I think more important to agree and move forward. So if option 2 above is chose, I would be happy.

marcusrbrown:
@avtarsehra I wanted to confirm that you were OK for me to post your ECIP as an EIP in the relay protection thread at ethereum/EIPs#134, and make sure that you didn't have any further changes to it before it was presented.

I like option 2 the best as IMO it provides the greatest protection at a fairly cheap price. It allows us to completely separate the chains (nonce and mod would, but there are still ways to circumvent) while remaining compatible with other proposed ECIPs/EIPs and fixes. Even with the extra field, the EVM, contract deployment, consensus, etc. all remain unchanged. It also allows us to clearly separate our own forks with certainty.

I think we should present your ECIP as is with an additional option to have it represented as an extra TX field instead of an extra hashing step.

avtarsehra:
@igetgames yes that sounds good, I'm ok with that. If you're going to submit in EIP, it would be best to include all as one post, so feel free to combine the way you think is most efficient and effective.

realcodywburns:
Will all existing contracts have to be redeployed?

realcodywburns:
No on second thought, the contracts still rely on underlying nodes which would get the updated tx protocols.

realcodywburns:
Can this topic be closed?

PR #16, opened 2016-10-10
Create ECIP-1013

realcodywburns:
Initial commit


splix:
I've an implementation of this as a smart contract - https://github.com/splix/toporick

PR #47, opened 2017-04-29
[WIP] Ethereum Classic Protocol Specification

sorpaas:
This ECIP proposes "Ethereum Classic Protocol Repository (ECPR)", consisting of documents of the Ethereum Classic protocol specifications. It also proposes to fork Ethereum's yellow paper as ECPR-0.

[Rendered ECIP](https://github.com/sorpaas/ECIPs/blob/master/ECIPs/ECIP-1020.md) and [rendered protocol specification documents](https://github.com/sorpaas/ECIPs/tree/master/Protocol).

Current TODOs:

* [ ] Allocate an ECIP number and change the currently fake one to the real one....
* [ ] Remove not-applied-to-all-block specifications from the yellow paper.
* [ ] Better formatting and explanation for individual additional specs.
* [ ] Check everything confines with the current implementations (Geth Classic and Parity).

Also, right now Github doesn't have math equation rendering support, so I suggest we create `gh-page` for this repo and rendering there instead.


sorpaas:
This is being replaced by https://github.com/ethereumproject/rfc `9/MAIN`. Let's move the discussion there. 😃

PR #66, opened 2017-06-29
ECIP-1025: Precompiled Contracts for zkSNARK Verification

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip1025/init/ECIPs/ECIP-1025.md))

**TLDR**, this allows certain anonymity within the Ethereum Classic network.

This combines [EIP-212](https://github.com/ethereum/EIPs/pull/212) and [EIP-213](https://github.com/ethereum/EIPs/pull/213).

zkSNARK verification will allow anonymous transaction to be executed on the Ethereum Classic network. See [this](https://github.com/zcash/babyzoe/blob/master/zoe/contract/mixer.sol) for how a simple mixer contract can be implemented using zkSNARK verification. This ECIP implements three primitive operations in order to perform zkSNARK verification. This allows changes of zkSNARK algorithms without requiring another hard fork.

The general benefit of zkSNARKs for Ethereum and Ethereum Classic is that it will increase the privacy for users (because of the Zero-Knowledge property) and might also be a scalability solution (because of the succinctness and efficient verifiability property).

### DIFF
--- a/ECIPs/ECIP-1025.md
+++ b/ECIPs/ECIP-1025.md
@@ -1,0 +1,168 @@
+## Title
+
+    ECIP: 1025
+    Title: Precompiled Contracts for zkSNARK Verification
+    Author: Wei Tang <hi@that.world>
+    Status: Draft
+    Type: Consensus (hard-fork)
+    Created: 2017/06/29
+
+<br />
+
+## Motivation
+
+Precompiled contracts for elliptic curve and pairing operations are required in order to perform zkSNARK verification within the block gas limit.
+
+## Abstract
+
+zkSNARK verification will allow anonymous transaction to be executed on the Ethereum Classic network. See [this](https://github.com/zcash/babyzoe/blob/master/zoe/contract/mixer.sol) for how a simple mixer contract can be implemented using zkSNARK verification. This ECIP implements three primitive operations in order to perform zkSNARK verification. This allows changes of zkSNARK algorithms without requiring another hard fork.
+
+The general benefit of zkSNARKs for Ethereum and Ethereum Classic is that it will increase the privacy for users (because of the Zero-Knowledge property) and might also be a scalability solution (because of the succinctness and efficient verifiability property).
+
+This combines [EIP-212](https://github.com/ethereum/EIPs/pull/212) and [EIP-213](https://github.com/ethereum/EIPs/pull/213).
+
+## Specification for Addition and Scalar Multiplication on the Elliptic Curve alt_bn128
+
+Add precompiled contracts for point addition (ADD)  and scalar multiplication (MUL) on the elliptic curve "alt_bn128".
+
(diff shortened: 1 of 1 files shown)


sorpaas:
Right now I'm in favor of creating pre-compiled contracts that allows implementing something similar to Monero's RingCT ([overview](https://getmonero.org/resources/moneropedia/ringCT.html), [technical document](https://lab.getmonero.org/pubs/MRL-0005.pdf)) rather than zkSNARK.

The problem with zkSNARK is that the setup phrase is too expensive. Zcash requires downloading more than 800MB of initial data for starting the process. Even we only include the minimal or reuse Zcash's variables, I still think it would cause some trouble when trying to put all those information on the blockchain.

At the same time, toxic waste, no matter how careful people are when generating it, might still be a trust issue.

tldr, I think even if ECIP-1025 is activated we would have trouble to actually use it. Alternative technology like RingCT should be analyzed before we make the decision.

sorpaas:
It turned out Ethereum has a ring signature implementation [a long time ago](https://github.com/ethereum/serpent/blob/develop/examples/ecc/ringsig.se). It uses the old Monero one but not RingCT, however.

For this ECIP, it would be great if we can ask/understand why @vbuterin switched from Monero-like anonymous transactions to Zcash-like ones.

realcodywburns:
[Here ya go](https://blog.ethereum.org/2016/01/15/privacy-on-the-blockchain/)  I prefer ring signatures because they are less computationally costly and they do not require the same level of trust as snarks. The disposal of zk toxic waste is an unresolved issue. For 'privacy coins' my preference has been monero over zcash.

realcodywburns:
It is worth mentioning ring signatures by themselves do not provide complete privacy.  Stealth addresses and confidential transactions are combined to obfuscate transactions details in monero. Stealth addresses are 'easy' enough to produce by adding `R` to the data field of a transaction and having a wallet that is capable of parsing transactions for such transactions( I may write an ecip for these in any case) and do not truly require any chain modification. Confidential transactions use Pedersen commitment schemes and ring signatures and might be able to be done in a smart contract, but it woild be costly.

realcodywburns:
ECIP-1035: ETC-PE Stealth Address #78

realcodywburns:
A new, more antifragile mpc process is planned for the next phase of zk. I am currently participating and will report my role after the conclusion of the "powers of tau" event. Also. Jpmchase quorum, a clone of geth for consortium chains has implemented zsl which are zk pre-compiles for shielding and unshelding transactions.

Mpc: https://github.com/ebfull/powersoftau
Zsl: https://github.com/jpmorganchase/zsl-q

arielgabizon:
@sorpaas note that only the verifying key needs to be on the blockchain which is a few hundred bytes. The big thing is the proving key.

PR #69, opened 2017-07-18
ECIP-1027/28: Scaling ETC with Sidechains [WIP]

splix:


ECIPs/ECIP-1027.md
```diff
@@ -0,0 +1,218 @@
...
+```
+
+## Geth Enhancements
+
+Geth should have support for multiple consensus mechanisms, and a configuration for multiple chains. Part of this 
+is already implemented, and we can have multiple custom/private chains with Geth starting at version 3.4.0. Similar
+features are already implemented in Parity.
+
+It will also require improvements in user interface, new commands, configuration options and documentation. None of 
+this is a breaking change
```
elaineo, line 209: It sounds like C1 and C2 are done as federated sidechains (like RSK on bitcoin - is that correct) but I'm not sure I understand C3. Does something comparable exist already?

splix: RSK is more C2 imho. Lighting/Raiden could be an example of C3

ECIPs/ECIP-1027.md
```diff
@@ -0,0 +1,218 @@
...
+Smart contracts for controlling shared state, and Transaction Replay like logic can give users additional security for 
+their cross-chain operations.
+
+_Example_: a smart contract based Token Exchange with Deposit/Withdraw through a special Smart Contract in Parent chain
+
+### Public Chain (C3)
+
+Public Chain is a general use blockchain, operated by multiple parties and in an untrusted environment. It can use a 
+fast PoS consensus algorithm for fast and cheap transactions, or individual configuration of EVM.
+
```
ghost, line 109: Is `C1/C2/C3` some common sidechain classification or was invented by you @splix?

ECIPs/ECIP-1027.md
```diff
@@ -0,0 +1,218 @@
...
+## Checkpoint authorization
+
+Making a checkpoint will require verification mechanism to avoid corrupted state and to protect from malicious actors.
+
+For simple cases, when Sidechain is solely managed by one business (C1/C2 types), it’s enough to authorize an address 
+(or set of addresses) which can make such checkpoints.
+
+For more sophisticated chains (C2/C3 types) verification a Sidechain can use methods similar to Proof Of Stake, or 
+even implement full verification in a smart contract. Later is possible if a checkpoint can provide security by 
+validating only a chain headers and part of shared state.
```
ghost, line 119: I find that section most interesting - really how to connect two blockchain together and to reach consensus, but you left technical details beyond the scope of the document. What about the next steps to work out some solution for this?

ECIPs/ECIP-1027.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Cross Transactions
+
+Cross-transactions for public Sidechain (C3) is most complicated part. In additional to untrusted environment it 
+can have multiple shared states managed by different types of services. It seems to be infeasible to implement a 
+general algorithm which will work for all cases.
```
ghost, line 125: Can you please add a definition for `cross-transaction`, what did you mean exactly? Multiple shared states are states from separated blockchains between which running cross-transactions?

ECIPs/ECIP-1027.md
splix: "invented here". just need something as a reference cross document

ECIPs/ECIP-1027.md
splix: There're few possible ways, so it's better to introduce them as separate ECIPs

ECIPs/ECIP-1027.md
splix: A state change on one chain resulting in a state change on another chain. This state modification can be different, but inseparable linked

RorschachRev:
Can we further leverage side chains to be an effective cross-chain technique? For example, a sidechain of BTC to ETC transactions with validation. Same for Zcash and Monero? (Where the privacy techniques make verification expensive in gas.) I think that would be considered a C3 with externalities?

hannibalhuang (commented):
Just leave a note here that I will work with ETCDEV team on implement a trial sidechain for OpenStack Public Cloud Passport Program :) Slides could be found: https://docs.google.com/presentation/d/1RYRq1YdYEoZ5KNKwlDDtnunMdoYRAHPjPslnng3VqcI/edit?usp=sharing

pyskell:
Are there any repos with sidechain code/examples besides the ECIPs? Does Geth have anything? I see the ECIP says there is some capability since geth 3.4.0.

phyro:
Are there any updates to the proposal? if yes please update it when you find the time, thanks! :+1:

gagarin55:
@splix, ECIP-1028 contains wrong ref to parity wiki, could you update it to https://wiki.parity.io/Proof-of-Authority-Chains,  please

PR #72, opened 2017-08-23
ECIP-1030: Defining the SYSTEM Transaction

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip-1030/ECIPs/ECIP-1030.md))

This is a "Meta/Informational" ECIP that aims to make it easy for us to discuss future related protocol upgrades, for example, EIP97.

### Abstract

This gives a standardized view of SYSTEM transactions -- transactions that are enforced by block validity rules (i.e. if the transaction is not included, it is considered invalid) but not executed by users. Note that this does not change consensus at all, but just gives a unified view of a special type of transactions (which we already have -- block rewards) in preparation of future extension (for example, EIP96 (https://github.com/ethereum/EIPs/pull/210) is one that uses SYSTEM transactions). Using this view, an Ethereum Virtual Machine itself can handle all account state changes without needing additional work from the client.

### DIFF
--- a/ECIPs/ECIP-1030.md
+++ b/ECIPs/ECIP-1030.md
@@ -32,7 +32,7 @@ This SYSTEM transaction would have its `gas_price` set to 0, and `gas_limit` set
 
 ### Implementation
 
-This ECIP has been implemented in SputnikVM (https://github.com/ethereumproject/sputnikvm).
+This ECIP has been implemented in SputnikVM (https://github.com/ethereumproject/sputnikvm/pull/251).
 
 ### Discussion and Concerns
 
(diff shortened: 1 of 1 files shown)


pyskell:
So I have a question, is SYSTEM intended to be just for storing state or also executable? (https://github.com/ethereum/EIPs/pull/210) seems to be storing state only

sorpaas:
@pyskell EIP97/EIP210 needs a transaction call at each block for storing state. So for them, it is also executable.

And in here, yes, it is executable. This just define a generic way to consider SYSTEM transactions. A SYSTEM transaction can do anything that a user-executed transaction can do, without costing balance or gas, but the actual thing they can do is restricted by the block validity rule.

splix (approved):

elaineo (approved):

sorpaas:
I'm going to merge this for ease of categorization.

Note that this ECIP will still stay in Draft, so please feel free to continue to review or submit PRs against this ECIP for changes.

PR #74, opened 2017-08-29
ECIP-1031: Recursive Length Prefix Media Type

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip1031/ECIPs/ECIP-1031.md))

### Abstract

Recursive Length Prefix (see [Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) Appendix B) is the data type used for all common Ethereum Classic structures including blocks, raw transactions, receipts and accounts. Defining a media type for Recursive Length Prefix (RLP) would allow a RESTful HTTP server to return raw RLP information directly.

### DIFF
--- a/ECIPs/ECIP-1031.md
+++ b/ECIPs/ECIP-1031.md
@@ -1,0 +1,29 @@
+### Title
+
+    ECIP: 1031
+    Title: Recursive Length Prefix Media Types
+    Author: Wei Tang <hi@that.world>
+    Status: Draft
+    Type: Standard
+    Layer: Meta
+    Created: 2017-08-29
+    
+### Abstract
+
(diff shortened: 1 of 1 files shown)


sorpaas:
I managed to get application/vnd.ecip.rlp registered in IANA's Media Types database. See https://www.iana.org/assignments/media-types/media-types.xhtml

So now this media type can be used without worrying about conflicts, I would suggest to merge this.

pyskell:
I'm fine with this since it simply allows clients to return RLP data but does not define a structure. 

That being said a standard structure/endpoints for RESTful HTTP data from a client should be defined in another ECIP and may be best finalized in tandem with this ECIP to help avoid client software creating different implementations.

RorschachRev:
Making RLP a media type sounds fine. I am guessing that the code in this case would be RPC and IPC requests for RLP data would have an additional return type: the media header wrapped around the RLP.

sorpaas:
See comment https://github.com/ethereumproject/ECIPs/pull/82#issuecomment-346282918

PR #75, opened 2017-09-02
ECIP-1032: Readable Address and Transaction Hash

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip-1032/ECIPs/ECIP-1032.md))

### Abstract

This defines convertion of address and transaction hash in Ethereum Classic network to readable strings, similar to how [Urbit](http://urbit.org/) handles its ship names. As an example, EthMine's coinbase address (0xdf7d7e053933b5cc24372f878c90e62dadad5d42) written in this scheme becomes `~hocwyt~misper~difmev~dabryd~livtep~dotpun~nibbex~nalbyn~pocmug~morden`.

This ECIP only proposes an optional implementation for wallets and other dapp interface, and does not change anything in the underlying protocol, or in JSON RPC.

### Motivation

Ethereum Classic's address is 160-bit, and transaction hash is 256-bit. They're currently only represented in a long hex string. This, when exchanged by users (for example, in situations where it needs to be transmitted by hand) can result in errors. This ECIP proposes a way to rewrite the 160-bit address or the 256-bit transaction hash in a human readable way similar to Urbit.


elaineo:
Just to clarify, you're not proposing to replace existing address/ tx hashes, right? This is more of a UI for wallet interfaces, like how a private key can be recorded as a sequence of words?

BTW i still can't remember my urbit ship name ;(

sorpaas:
Yes you're right! @elaineo

arvicco:
It looks like Urbit base-65536 phonetic names, but uses different phonemes, right? Is there a reason for this? If it's going to copy Urbit's approach, why not use the same phonemes as well?

sorpaas:
@arvicco I intended to use the same phonemes as Urbit, but might have accidentally got an outdated one.

Do you know where I can find the newest definition of phonemes?

sorpaas:
Just fixed the phonetic base, should look much nicer now.

arvicco:
Yes, the phonetic table looks up to date with Urbit now.

arvicco:
Personally, I love the pure geekery of Urbit phonetics but in terms of declared goal of making obscure numbers memorable, I'm highly skeptical. Something like [this](https://github.com/paulbellamy/base65536) goes much longer in reaching this goal IMHO.

sorpaas:
Also probably something like https://github.com/ethereumproject/ECIPs/pull/76, if the goal is to be momerable. 😃

Indeed it doesn't help in terms of memorable. I think Urbit and this ECIP only makes it readable and pronounceable. It's usefulness depends how it's actually used. For example, I would be more willing to put a Urbit-like address in my personal website. Also, in the case that I need to copy/paste an address from a non-connected mobile phone to my laptop (in the case of type and looking), this scheme will also help.

sjmackenzie:
This is a seriously bad idea. It messes with the semantics, it'll confuse users.

sorpaas:
@sjmackenzie this ECIP does not change any existing semantics, rather it crates an optional new one that people can choose to use or not use. Probably you misread something?

PR #76, opened 2017-09-02
ECIP-1033: Mnemonic code for public address

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip-1033/ECIPs/ECIP-1033.md))

### Abstract

This extends [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) for public addresses on the Ethereum Classic network. As an example, address `0xdf7d7e053933b5cc24372f878c90e62dadad5d42` becomes `term-typical-life-include-deputy-toward-move-tourist-maple-goose-define-forget-suspect-push-any`.


arvicco:
You know what would be cool? If we could use not just words but random sentences following SOV structure. Like typical-goose-moves-irrelevant-deputy. 8-) 3 sentences like that would be quite memorable.

sorpaas:
I think that's doable, given we can make it "adj. n. v. adj. n." pattern. So like you said, it would be exactly three sentences. I think the only thing needed is to find 4096 good adj.s, 4096 good n.s, and 2048 good v.s. And it might also be only limited to English.

elaineo:
It's worth reviewing the BIP0039 criticisms -- lack of versioning, wordlist dependency.
https://github.com/bitcoin/bips/wiki/Comments:BIP-0039

Also I'm not sure that these mnemonic addresses address a real problem -- shouldn't we be encouraging users *not* to reuse addresses? Contracts might be static, but isn't it easier to verify the correct address with ECNS or an identicon or something?

**edit:** that said, i don't have a strong opinion for or against ECIP-1033 or ECIP-1032. Obviously it doesn't make sense to keep both.

sorpaas:
I think this ECIP is affected less by the BIP-39 criticisms. We're using only the first part of BIP-39 -- convert from or to mnemonic codes into a bytearray. That said, those criticisms still hold -- checksums still depend on wordlist, and version bits not existing might still cause trouble.

I agree with @elaineo that mnemonic address might not actually needed. But I do think something that is **readable** (ECIP-1032) would help people like me who have a set up that sometimes require hand copying an address from one device to another. Several parts of ECIP-1032 might be too geeky, but personally I still think the section "Hex Readability Fix" of ECIP-1032 might be counted as a small improvement.

PR #77, opened 2017-09-05
ECIP-1034: Message Signing and Verification in JSON RPC

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip-1034/ECIPs/ECIP-1034.md))

### Motivation

The current `eth_sign` definition is different across different client implementation, which creates [many](https://github.com/ethereumproject/go-ethereum/issues/323) [problems](https://github.com/paritytech/parity/issues/5490). In addition, there is currently no way to verify a signed message using JSON RPC. This ECIP defines two new RPC calls `sign` and `recover` in a different namespace.

### Namespace

The author of this ECIP recommends to define the two new RPC in a different namespace rather than `eth`. Reusing `eth` may cause confusion with the existing `eth_sign` RPC call and might further pollute the already huge namespace. We can use `etc` namespace, or `key` namespace (so those two RPC method becomes `etc_sign`, `etc_recover`, or `key_sign`, `key_recover`).

### DIFF
--- a/ECIPs/ECIP-1034.md
+++ b/ECIPs/ECIP-1034.md
@@ -1,0 +1,57 @@
+### Title
+
+    ECIP: 1034
+    Title: Message Signing and Verification in JSON RPC
+    Author: Wei Tang <hi@that.world>
+    Status: Draft
+    Type: Standard
+    Layer: JSON RPC
+    Created: 2017-09-05
+   
+### Motivation
+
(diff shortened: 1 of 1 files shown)


elichai:
Are there any news in this subject?, 
signing verifying should seriously be standardized (e.g removing the weird "Ethereum Signed Message" from some clients).
Not only this ecdsa+keccak256 isn't standard and is incompatible with some EC libraries, this at least should have a real doc saying how the sig should work and behave.

Issue #63, opened 2017-06-28
ECIP-?: Extension of ecip 1010 timeline


realcodywburns:
## Abstract

The original ECIP 1010 ( #4 ) describes the details of delaying Difficulty Bomb growth for one year. It was suggested to change Difficulty Bomb formula to stop growing for period of time from January 2017 to December 2017, which is around blocks 3,000,000 and 5,000,000 accordingly.  This would correspond with the end of era 1 per ECIP 1017 ( #15   ) and implementation of the new monetary policy. 

## History

The bomb was originally delayed to review: 

- Monetary Policy
- PoW/PoS or Hybrid
- Replay Attack
- DAG growth 
- Difficulty Bomb

 
Several of these items have been addressed, several more have been added in the past year including evm evaluation, file storage, and sharding. 

 ## ECIP
 
I propose  remaining at the current difficulty, extra difficulty 2**28, until block `9,000,000`  or render safe forever (setting the default unfreeze block to `0x7fffffffffffffff`)  as part of the activation of ECIP 1017. 

## Activation

This ECIP should be activated with an 'opt-out' flag. User wishing to keep the  current ecip-1010 deadline should be allowed the option to use a `--diehard` flag which will keep the bomb reactivation at block 5m. the default action (no flag) will be to move the bomb block to 9m

sorpaas:
I agree. Right now the protocol works well for us. Setting it to `0x7fffffffffffffff` would allow us to completely eliminate the difficulty bomb problem as we're staying PoW, and so we can focus on other much more important work.

sorpaas:
@realcodywburns Probably you meant that this should be as part of the activation of ECIP-1017 (monetary policy) as ECIP-1015 (the one about long-term gas cost changes) has already been activated?

realcodywburns:
correct 1017- monetary policy - edited to correct

elaineo:
I'm in favor of removing the bomb forever.

splix:
I'm in favor of removing it too. 

We may want to make another protocol upgrade after block 5M. That bomb will motivate to do it, say at block 6M. At this block we can do upgrades and remove bomb finally.

drd34d:
I'm in favor of removing the diff bomb too.

OmniEdge:
I am in favour to remove the bomb forever too.

dollebrekel:
I am in favor of removing the bomb.

carlox757:
I am in favor of removing the bomb.

[...]

PR #80, opened 2017-09-06
ECIP-1036: Fallback Complete Difficulty Bomb Diffusion

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip-1036/ECIPs/ECIP-1036.md))

### Abstract

This ECIP describes a fallback plan if an extension of ECIP-1010 does not activate. If that is the case, then completely diffuse the difficulty bomb at block 5452163.
    
### Motivation

There is currently a small probability that [an extension of ECIP-1010](https://github.com/ethereumproject/ECIPs/issues/63) might not be able to be activated. The probability is really slight, but it is still possible. Please see the discussion of that issue for reasonings from different parties. This ECIP provides a fallback plan, to give assurance for those who might be impantient or worried of the difficulty bomb, so as to avoid a possible community split or an unexcepted hard fork.

Block 5452163 is chosen as the fork block number. This is a totally arbitrary and meaningless number. In this way, we avoid collisions of the fork block number with other protocol upgrade hard forks.

### DIFF
--- a/ECIPs/ECIP-1036.md
+++ b/ECIPs/ECIP-1036.md
@@ -16,7 +16,7 @@ This ECIP describes a fallback plan if an extension of ECIP-1010 does not activa
 
 There is currently a small probability that [an extension of ECIP-1010](https://github.com/ethereumproject/ECIPs/issues/63) might not be able to be activated. The probability is really slight, but it is still possible. Please see the discussion of that issue for reasonings from different parties. This ECIP provides a fallback plan, to give assurance for those who might be impantient or worried of the difficulty bomb, so as to avoid a possible community split or an unexcepted hard fork.
 
-Block 5452163 is chosen as the fork block number. This is a totally arbitrary and meaningless number. In this way, we avoid collisions of the fork block number with other protocol improvement hard forks.
+Block 5452163 is chosen as the fork block number. This is a totally arbitrary and meaningless number. In this way, we avoid collisions of the fork block number with other protocol upgrade hard forks.
 
 ### Specification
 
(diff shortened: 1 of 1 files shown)


splix (approved):

whilei:
What about using a smaller block number instead, like `5123456`. I wanted to choose the smallest, and nearest, possible block number that can be achieved across human consensus. 

Given @splix's expected difficulty from https://github.com/ethereumproject/ECIPs/blob/master/ECIPs/ECIP-1010.md, we can expect to see difficulty raise ~quickly~ exponentially after the "explosion."
```
With this changes min difficulty value will become:

Block 2,500,000 == 2**23 == 8,388,608
Block 3,000,000 == 2**28 == 268,435,456
Block 4,000,000 == 2**28 == 268,435,456
Block 5,000,000 == 2**28 == 268,435,456
Block 5,200,000 == 2**30 == 1 TH
Block 6,000,000 == 2**38 == 274 TH
```

sorpaas:
Difficulty bomb increases x2 every 100,000 blocks. So if at block 5,200,000 it is 1GH, then at block 5452163 it would be at most 8GH. If we can still have 8TH/s hashrate (which is our current, but we should question this assumption), then it would mean less than 1s block time increase.

Just for reference, 5452163 is around March or April, three months after ECIP1017. If it is done too early (5123456 is around January or February), miners and users might have problem upgrading their clients.

whilei:
I'm for it; and the sooner these changes are included in code+releases the better, in my opinion. 

Proposed changes implemented in geth classic PR here: https://github.com/ethereumproject/go-ethereum/pull/409.

sorpaas:
Looks like we're going to use ECIP-1041 (#87) instead. Closing this.

PR #81, opened 2017-09-10
ECIP 1037: Simple non-interactive URI Scheme

realcodywburns:
Abstract

This ECIP proposes a URI scheme for interacting with ethereum contract and accounts in a standard format.

Motivation

The purpose of this URI scheme is to enable users to easily make payments and interact with smart contracts by simply clicking links on webpages or scanning QR Codes.

### DIFF
--- a/ECIPs/ECIP-1037.mediawiki
+++ b/ECIPs/ECIP-1037.mediawiki
@@ -77,7 +77,7 @@ For example, so long as the majority of users work in ETC units, values should a
 
 === Bytecode Transactions ===
 
-Events and functions in evm contracts are can be preformatted to know functions in the underlying contract. The first four bytes of the call data for a function call specifies the function to be called. It is the first (left, high-order in big-endian) four bytes of the Keccak (SHA-3) hash of the signature of the function. This information along with the appropriate hex encoded data to pass.
+Events and functions in evm contracts are / can be preformatted to know functions in the underlying contract. The first four bytes of the call data for a function call specifies the function to be called. It is the first (left, high-order in big-endian) four bytes of the Keccak (SHA-3) hash of the signature of the function. This information along, with the appropriate hex encoded data to pass, form the bytecode.
 
 ====Example====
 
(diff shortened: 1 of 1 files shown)


realcodywburns:
Changing `chainid` to `id` for brevity. Addational space can also be saved by using iban address vs account hex.

drd34d:
sounds like an awesome idea to make interaction stupidily simple!

ECIPs/ECIP-1037.mediawiki
```diff
@@ -0,0 +1,196 @@
...
+Please see the BNF grammar above for the normative syntax.
+
+[foo] means optional, &lt;bar&gt; are placeholders
+
+ <nowiki>ethereum:<address>&id=<chainid>[?amount=<amount>][?gas=<startGas>][?code=<data>][?label=<label>][?message=<message>]</nowiki>
+
+=== Examples ===
+
+Just the address:
+ ethereum:0xCe5ED529977b08f87CBc207ebC216859820461eE&id=61
```
elaineo: How about `chainId` instead of just `id`? to avoid confusion with `networkId`

elaineo:
@realcodywburns actually i think it should remain `chainid` to avoid confusion with network id.

realcodywburns:
What about `v` or `cid`?

Dexaran:
@realcodywburns 

```js
 function getETC(){
    require(block.number-payoutDrip >= payoutRate);
    msg.sender.transfer(dropValue);   
    }
```

>
>====Example====
>
>Contract C is at 0xdeadbeef... and has the following psudo function:
>
>Contract C requires no inputs and only requires enough gas to preform its functions. The byte code for >this function can be derived using `web3` as:
> 
>web3.sha3("getETC()");
>
>which yeilds:
>
>0x023b88685d10d29e0bf0563e4ab1b9d8fc8333c166705be5538c4b079cdd9af0
>
>The information that will be passed in the byte code field would then be:
>
>code=0x023b88685d

I think that if you need to pass the function signature then it should be `0x023b8868` but not 0x023b8868~~5d~~ .

realcodywburns:
I updated, it should only be the left most 4 bytes

realcodywburns:
https://github.com/ethereum/EIPs/issues/67

elaineo:
@realcodywburns i still prefer chainId for consistency, but it's not a dealbreaker or anything :)

realcodywburns:

PR #82, opened 2017-11-03
ECIP-1038: Jets in Ethereum Virtual Machine

sorpaas:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip-1038/ECIPs/ECIP-1038.md))

### Abstract

Jets are optimization marks. Using jets, users can place optmization marks in an EVM program. An EVM interpreter will then verify the optimization mark claim, and if the verification passes, execute a native code version of the program that is equivalent to the original program.

### DIFF
--- a/ECIPs/ECIP-1038.md
+++ b/ECIPs/ECIP-1038.md
@@ -5,7 +5,7 @@
     Author: Wei Tang <hi@that.world>
     Status: Draft
     Type: Standard
-    Layer: Consensus
+    Layer: Application
     Created: 2017-11-04
     
 ### Abstract
--- a/ECIPs/ECIP-1038.md
+++ b/ECIPs/ECIP-1038.md
@@ -43,8 +43,10 @@ Usually for an opcode sequence to be replacable by native codes, it is required 
 Below we consider the opcode `MUL` does not exist, and implement a jet-able version of `MUL` using `ADD`:
 
 ```
-DUP1
-DUP3
+PUSH1 00
+DUP2
+DUP4
+ADD
 ADD
 SWAP2
 PUSH1 01
@@ -53,7 +55,7 @@ SUB
 SWAP2
 DUP3
 PC
-PUSH1 08
+PUSH1 11
 SUB
 SWAP1
 JUMPI
@@ -61,3 +63,5 @@ SWAP2
 POP
 POP
 ```
+
+This can be simplified by `MUL` -- taking two items from the stack, and push one new item which is the multiply of the two items. Its gas cost can also be simplified as `12 + 51 * <mul second item>`.
--- a/ECIPs/ECIP-1038.md
+++ b/ECIPs/ECIP-1038.md
@@ -1,0 +1,63 @@
+### Title
+
+    ECIP: 1038
+    Title: Jets in Ethereum Virtual Machine
+    Author: Wei Tang <hi@that.world>
+    Status: Draft
+    Type: Standard
+    Layer: Consensus
+    Created: 2017-11-04
+    
+### Abstract
+
+Jets are optimization marks. Using jets, users can place optmization marks in an EVM program. An EVM interpreter will then verify the optimization mark claim, and if the verification passes, execute a native code version of the program that is equivalent to the original program.
+
+### Specification
+
+The optimization mark is defined as two adjencent opcodes.
+
+```
+PUSH11 45 56 4d 4a 45 54 53 <4 bytes of registry ID>
+POP
+```
+
+Readers can immediately follow that this optimization mark does not have side effects other than costing 5 gases. `45 56 4d 4a 45 54 53` is a magic code. The `<4 bytes of registry ID>` is interpreted as a big-endian unsigned integer.
+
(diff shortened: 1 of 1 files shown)


splix:
I have a question: if VM already has a db of native code and can verify against it, why a developer need to add a special mark to force it? can't it be done automatically w/o external request?

sorpaas:
If what you are talking about is account `code_hash` to native code mapping, then yes. That's the extended version of precompiled contracts, and SputnikVM already supports it. (To add this to Parity or Geth is also not a hard thing.)

However, many contract only differs for a subset of its code, and to optimize them, we want to replace a subset (and commonly occurring) code to native one. This cannot be done without marking, because otherwise the complexity would be `O(nm)`, meaning that you need to match each line (`n`) repeatedly with each item (`m`) in the existing database.

splix:
I don't like idea of having a hack to let VM know more that it can learn from bytecode itself. It's compiler job to optimize for evm, and put everything in order that can be executed in optimal way by VM (i.e. it should able to find a pattern automatically). 
I'm in favor what JVM JIT does, starting from basic https://docs.oracle.com/cd/E13150_01/jrockit_jvm/jrockit/geninfo/diagnos/underst_jit.html and there're tons of researches that we can apply to out VM

RorschachRev:
The problem I see with this is that it breaks all compatibility. If this ECIP is adopted, it should be bundled with additional opcodes or other EVM breaking functionality. The obvious advantage here is that everyone who adopts Jets can process code chunks faster. If we can make Jets an "optional" instruction that is skipped by non-compliant clients, that is great. We would offer performance gain with no downside. 

I'd like to know more about verification, and where the Jet execution code would be located. Are we going to lower gas costs on Jet code, or go with the "native EVM cost" and only use Jets to speed execution?

I think Jets sidestep the issue of "needing 30 new OPCodes" very cleanly, and some code that currently takes 3 million gas could be done with execution costs approaching 6k gas. Are we measuring execution time of a Jet for accurate gas accounting in the future?

This is a huge topic, and warrants more thought.

sorpaas:
@RorschachRev No, jets doesn't do what you said. I think there's some misunderstanding.

Jet marks are **normal opcodes in the current consensus rule**. So it does not break compatibility. Existing EVM runs jet marks just fine, just being slower without the native code. Jets requires that the native code that it replaces cost the same gas, and have the same side effect. So there's no "native EVM cost" or anything like that. The cost is the same.

I think what you are talking about is more like pre-compiled contracts. That's a separate topic and there's indeed something to be reviewed as well: https://that.world/~essay/pdf/non-hard-fork-precompiled.pdf

RorschachRev:
splix documented something like Pysco for speeding up python programs at run or compile time. That work was one of the early attempts at supercompilers, and it worked well for non-optimized code. Adding this functionality inside solc and other compilers makes sense, I don't see any reason why this should be in the EVM.

@sorpaas I thought Jets were like the paper you linked, but sidestepping the EVM into C/OpenCL. Either branch would be valid, but the cost of doing the same operations in EVM would be 3 mil gas and doing the operations in C/OpenCL would be 10k gas.  Doing it with a contract in your paper still requires a hard fork to create a separate contract block and a new network message right? After that infrastructure is in place, you have a block of contracts that are similar to Oracles in that they compute based on input, and then generate an in blockchain answer that the nodes can process. Why not skip the whole hard fork, skip the dedicated block and the new message architecture and just create a series of "compute Oracles"? If the market adopts 5 of those computations, we could put the popular opcodes in the EVM, with a fallback to use the compute oracle. That way no hard fork is required, and the work could be done by oracle or by miner. Paying the Oracle(s) (I like 3 oracles) would be done "per problem" x "number of users" which allows smooth migration away from the oracle. Actually allocating the ETC vs Gas is something I'd have to think on more, but you may have a better solution.

I don't think the EVM should be extended with Jets, I think it should be done by compilers.

sorpaas:
(@RorschachRev I still think we're talking about different things. This ECIP was about an EVM implementation-specific details and has nothing related to hard forks, but let's continue the discussion later.)

I have a feeling that we tries to mix two different types of specifications here and thus make several stuff hard to move forward:

* **Consensus-related specifications**: Those are specifications about hard/soft-fork and the Ethereum protocol at large. This is the place where we have `ACK`s and `NACK`s and even many political discussions. It is comfortable in those specifications to get blocked due to big or small reasons because after all, it is related to consensus.
* **Non-consensus-related specifications**: Those are specifications that is only related to anyone who implements this specifications. Troubles would only go to them. Those specifications are published for standardization. Users would not be bothered whether an EVM has jet implemented or not, but implementors would be bothered if there're more than one EVM jet standards. So the focus is not on **whether it should** get implemented, but **how it should** get implemented.

So, as an experiment, announce [ethoxy/specs](https://github.com/ethoxy/specs), a specification repository that *only accepts non-consensus-layer specifications*. The repo has maturity level in replace of ECIP's status. And instead of "reaching consensus", it is defined in terms of whether an implementation has come in place. This would make implementor's life a lot easier.

In the beginning, I'm going to move this, together with the RLP media type ECIP to that repo and see how it goes...

Issue #15, opened 2016-10-06
ECIP-1017 ETC monetary policy and emission schedule


arvicco:
Some thoughts on ETC Monetary Policy, copied from https://ethereumclassic.github.io/assets/Ethereum_Classic_-_The_New_Original_Innovator.pdf

Platform token aligns economic incentives of users, miners and investors
Current Ethereum monetary policy undefined (unlimited token inflation)
ETH policy is supposed to change with a transition to PoS unpredictably
Lack of token scarcity does not support long-term investor confidence
ETC transition to a long-term PoW necessitates monetary policy decision
ETC commitment to a specific monetary policy:
provide certainty to the market
boost investor confidence
creates competitive differentiator to ETH with its undefined policies
Fixing ETC Monetary Policy - Options:

Keep unlimited token inflation forever, like current ETH

Pros: keeps the status quo, encourages token spending
Cons: long-term monetization uncertain, not a good store of value/investment
Fixed final supply with Bitcoin-like reward cut-offs (halvings)

Pros: experimentally proven to work for BTC, sustainable token monetization due to scarcity
Cons: changes existing inflationary parameters, disruptive “halving” events
Fixed final supply with exponential reward adjustments (per epoch)

Pros: no disruptive “halvings”, sustainable token monetization due to scarcity
Cons: changes existing inflationary parameters

---

Slack:@snaproll (sorry, can't find you on Github) started exploring ETC supply model here:
https://docs.google.com/presentation/d/1jWTiJr7FmGjTyNj80-OapxDOiRGKVwmlkQ63LOnPO24/

arvicco:
Log of 2016-10-28 Monetary Policy discussion from Slack#development: 
https://docs.google.com/document/d/1jVM3iN1ClVrDjZ4AcJRAQyn24B1NN6Y5_X5IpLrT9Ss

mikeyb:
Currently a [Pull Request](https://github.com/ethereumproject/ECIPs/pull/20) for ECIP-1017.

Reading through the details and will add my comments/concerns as necessary.

realcodywburns:
I am adding a reference to this issue on the published ecip, I plan to close the issue, if there is no objection, on block 3,277,945.

snaproII:
Thesis uploaded to GitHub for easier retrieval, if needed: 

https://github.com/snaproII/Crypto/blob/master/ECIP-1017/Thesis.md

snaproII:
[...]

PR #83, opened 2017-11-16
ECIP-1039: Monetary policy rounding specification

whilei:
([Rendered](https://github.com/ethereumproject/ECIPs/blob/ecip-1039/ECIPs/ECIP-1039.md))

### Abstract

This ECIP proposes a precise specification for eventual rounding issues around
[ECIP1017 Monetary Policy](https://github.com/ethereumproject/ECIPs/blob/master/ECIPs/ECIP-1017.md) calculation, particularly around

- block winner reward calculation
- block winner reward for including uncles
- block uncle miner reward for having uncles included in the winning block

The content of the proposal deals exclusively with reward calculation beginning with Era 2.

cc @rtkaczyk, @snaproII 

Rel #15, https://github.com/ethereumproject/go-ethereum/issues/352


ECIPs/ECIP-1039.md
```diff
@@ -0,0 +1,649 @@
...
+
+#### __Block winner reward for including uncle(s).__
+
+- Uncle(s) inclusion reward should be calculated and rounded down prior to addition
+  with the block winner reward.
+
+_eg._
+```
+blockWinnerReward = blockWinnerRewardAtEra(n) # Block winner reward rounds down once prior to division
+uncleInclusionReward = floor(1/32 * blockWinnerReward) # Block uncle inclusion reward rounds down per uncle
```
sorpaas: I think it's better to change this to `blockWinnerReward / 32` and then define `/` as "floor divide" because this symbol is also used above. Besides, using `1/32` might cause confusion of whether the calculation is done in floating points.

sorpaas (dismissed):
LGTM. Also added the PR to Parity. Please help review that.

shanejonas (dismissed):

shanejonas (approved):

ECIPs/ECIP-1039.md
whilei: Changed in 537451d

rtkaczyk:
LGTM

realcodywburns:
Ack 🖒

whilei:
relevant code links for reference
- PR to Parity: https://github.com/paritytech/parity/pull/7067
- in geth: https://github.com/ethereumproject/go-ethereum/blob/master/core/state_processor.go#136

rtkaczyk:
Also: [mantis](https://github.com/input-output-hk/mantis/blob/master/src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala)

whilei:
Well it sounds like this is a go. Final call for criticism or refinements... t-minus 24 hours and I'll merge it.

Dexaran:
Alright.

We have successfully implemented the monetary policy and it's the time for me to ask the question: **what is the main goal of this fixed-supply monetary policy?**

As I can see at the [London presentation](https://github.com/snaproII/Crypto/blob/master/ECIP-1017/London_Presentation.md) the main goal of this monetary policy is  (1) to attract investors, (2) boost the price and increase a network effect of ETC chain thus boosting the adoption, (3) establish a MP which sets out to achieve "optimal total investment". Is it so or am I missing something?

whilei:
@Dexaran I'm not sure what the goal of your question is. 

There are a lot of resources available around informing the monetary policy specification and reasoning, which have been around for ~8 months, including and beyond that one presentation.  

Or are you looking for a single thesis statement you can point specifically to for some reason?  Or are you trying to begin a discussion about it?

Dexaran:
I'm looking for a kind of single thesis explanation about it so that I can refer to. I'm asking here because I'm looking for an answer from those who decided to propose this changes.

As for me, it seems to be obvious that changing a monetary policy from "unlimited supply" to "capped supply" will cause a price increase. It's just a straight logic and a simple math.

The question is: what is the purpose of this? As for me it seems that the answers are (1) to attract investors and (2) increase network effect.

But I am not the one who proposed this changes. And now I'd like to know if I am missing the point or no. 

I'm not looking for opportunity to start a discussion because this makes no sense. It is already implemented and I don't see any value in rolling it back. A simple answer would be sufficient.

whilei:
Oh, OK. Then yea, I think you got the right idea. Supply, demand, value. 

But as for "purpose," I think you might be better off thinking of "purposes," since the final policy was a result of a lot of people's input, compromises, and cooperative decision making.  There are purposes, causes, and effects for stakeholders, gladhanders, theorists, enthusiasts, hobbyists, hecklers... 

Nor am I the one who proposed the changes, but - and if I may go so far - I've heard @snaproll once say the purpose of _his_ proposal was to "help find the best solution to an annoying problem so we can get the fuck on with our lives."  

And if I've misquoted him, I'll gladly claim the sentiment instead for myself ;)

snaproII:
@whilei is pretty close. The goal for me was to "remove the variable, so we can re-focus on utility." That was the initial driver for attacking this problem. If the data/arguments suggested that keeping the original emission rates was just as good, I think everyone would have been fine with that. But it turned out 5M20 was better.

snaproII:
Also, @Dexaran, "removing the variable" also means to minimize hesitation and skepticism. The crypto space is full of scam coins with wacky emission rates, emission adjustments made on a whim, premines, and the most insane concept ever: coins with the majority of total emission being from the premine that then move/propose a move to Proof of Stake. This is all noise that creates complexity without any utility. By attempting to mirror bitcoin's emission rate, the most well known emission rate, the "variable" of Monetary Policy is removed from ETC, and we can re-focus on if this decentralized script crypto can actually change the world or not - we all think it is interesting, but there still is nothing created that you can point to and say "LOOK, LOOK! THATS THE SMART CONTRACT KILLER APP!"

PR #96, opened 2018-07-27
Ethereum Non-fungible token standard

dpdanpittman:
I would like to request the erc721 nft standard be implemented.

### DIFF
--- a/ECIPs/ECIP-1042.md
+++ b/ECIPs/ECIP-1042.md
@@ -1,5 +1,5 @@
 
-### What is an ECIP?
+### Title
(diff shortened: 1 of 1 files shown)


splix:
If I understand correctly ERC-721 is only a smart contract interface, i.e. it's not a change of protocol, EVM, etc. If I'm right, there is no problem with using it as is, without having a ECIP document, same way as ERC-20 is used.

dpdanpittman:
@splix thank you for your response.  This is the way I see it as well I will use it as is!

PR #98, opened 2018-12-04
FISSION Status Codes

bmann:
Hello Ethereum Classic Team!

We've been speaking with @pyskell and others who encouraged us to submit our FISSION Status Codes so that it can be used in Ethereum Classic. If there is interest, we would then also submit FISSION Translate, which builds on chain translation for smart contract messages.

This is meant to be blockchain agnostic, as bytes are able to be coded in pretty much any system. This is standards track on Ethereum as ERC-1066, and various other standards are starting to use it as a dependency.

On Ethereum Classic, it looks like it would make the most sense as an Informational ECIP -- please let me know if you would prefer it be Standards track instead.

@expede is the author and architect of this work, I'm helping to learn the ECIP process and help explain anything that people have questions about.

The FISSION Suite has its own website, and the web version of this proposal may be more readable there: https://fission.codes/fission-codes/


realcodywburns:
Adding as ECIP-1050 @ https://github.com/ethereumclassic/ECIPS/ECIPS/ECIP-1050.md

